PR #13, opened 2020-03-12
Update filtered number in docs

akrabat:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Update the code example for converting a formatted number to a double to
match reality when the code is run.

See this [test](https://github.com/laminas/laminas-i18n/blob/master/test/Filter/NumberFormatTest.php#L121-L122).


michalbundyra:
Thanks, @akrabat!

PR #16, opened 2020-03-14
Adds new documentation for PostCode validator

froschdesign:
|    Q          |   A
|-------------- | ------
| Documentation | yes


docs/book/validators/post-code.md
```diff
@@ -0,0 +1,162 @@
...
+```php
+$validator = new Laminas\I18n\Validator\PostCode(['locale' => 'de_AT']);
+
+echo $validator->getLocale(); // 'de_AT'
+```
+
+### Default Value
+
+By default, if no locale is provided, `PostCode` will use the system locale
+provide by PHP's `Locale::getDefault()` and `Locale::getRegion()` to extract
```
michalbundyra: "provided by PHP's" <- missing "d"

docs/book/validators/post-code.md
```diff
@@ -0,0 +1,162 @@
...
+the region code.
+
+## Using Custom Format
+
+Postal code formats are regular expression strings. When the international
+postal code format, which is used by setting the locale, does not fit your
+needs, then you can also manually set a format by calling `setFormat()`.
+
+
```
michalbundyra: one empty line, please

docs/book/validators/post-code.md
```diff
@@ -0,0 +1,162 @@
...
+$validator = new Laminas\I18n\Validator\PostCode();
+$validator->setFormat('AT-\d{4}');
+
+var_dump($validator->isValid('AT-1010')); // true
+```
+
+### Conventions for self defined Formats
+
+When using self defined formats, you should omit the regex delimiters and
+anchors (`'/^'` and  `'$/'`). They are attached automatically.
```
michalbundyra: As I have checked in the code it is true when we skip delimiters, completely, but if we use other delimiters than `/` it will break the regexp by attaching additional `/^` and `$/`. We can provide regexp as: `#^AT-\d{3}$#` but internally it will be broken and changed to: `/^#^AT-\d{3}$#$/`

docs/book/validators/post-code.md
```diff
@@ -0,0 +1,162 @@
...
+class.
+
+Internally, the additional validation is done by
+[laminas-validator's `Callback`](https://docs.laminas.dev/laminas-validator/validators/callback/)
+class.
+
+```php fct_label="Constructor Usage"
+$validator = new Laminas\I18n\Validator\PostCode(
+    [
+        'service' => function ($value) {
```
michalbundyra: Can we use please static closure in our examples?
I think it is a good practice and we should promote it across whole documentation.

docs/book/validators/post-code.md
froschdesign: The question that arises here: is the documentation wrong or the code which handles this behaviour?
In my opinion the simple handling should be in the foreground and the code should realize this.

michalbundyra: I would prefer to detect any delimiter used, but the question is if we can do that in any reliable way?
Hm... I would need to think about it.

In general I do not like idea of attaching `/^` and `$/` (maybe `/` is the most used, but is not the only/valid one and there are sometimes good reasons to use another).

froschdesign: If we have a problem here then a separate issue report is needed.

michalbundyra: @froschdesign I've created issue: https://github.com/laminas/laminas-i18n/issues/18

michalbundyra:
Thanks, @froschdesign!

PR #17, opened 2020-03-15
Adds new documentation for IsInt validator

froschdesign:
|    Q          |   A
|-------------- | ------
| Documentation | yes


docs/book/validators/is-int.md
```diff
@@ -0,0 +1,111 @@
...
+
+Integer values are often written differently based on country or region. For
+example, using English, you may write `1234` or `1,234`; both are integer
+values, but the grouping is optional. In German, you'd write `1.234`, and in
+French, `1 234`.
+
+`Laminas\I18n\Validator\IsInt` will use a provided locale when evaluating the
+validity of an integer value. In such cases, it doesn't simply strip the
+separator, but instead validates that the correct separator as defined by the
+locale is used.
```
michalbundyra: I might be wrong, but something sounds to be wrong for me: "as defined by used locale"?

docs/book/validators/is-int.md
```diff
@@ -0,0 +1,111 @@
...
+```php
+$validator = new Laminas\I18n\Validator\IsInt();
+
+$validator->isValid(1234);    // returns true
+$validator->isValid(1234.5);  // returns false
+$validator->isValid('1,234'); // returns true
+```
+
+By default, if no locale is provided, `IsInt` will use the system locale
+provide by PHP's `Locale` class and the `getDefault()` method.
```
michalbundyra: "provided by PHP's" - missing "d" in "provided"

michalbundyra:
Thanks, @froschdesign!

froschdesign:
@michalbundyra 
Thanks for your quick help and review on Sunday! 👍

PR #21, opened 2020-03-16
Adds new page for migration to documentation

froschdesign:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Enhancement        | yes

### DIFF
--- a/docs/book/migration/migration-to-v2.4.md
+++ b/docs/book/migration/migration-to-v2.4.md
@@ -1,0 +1,31 @@
+# Migration to Version 2.4
+
+## General
+
+Version 2.4 adds support for PHP 7.
+
+## Validators
+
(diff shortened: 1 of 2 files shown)


docs/book/migration/migration-to-v2.4.md
```diff
@@ -0,0 +1,31 @@
+# Migration to Version 2.4
```
michalbundyra: I am not sure if it is the best header, maybe I am nitpicking here.

My thinking is that some people might migrate from 2.2 to 2.8 and they should be aware of it.
Maybe worth changing to "Migration from versions prior to 2.4" or something similar?

michalbundyra:
Thanks, @froschdesign!

PR #22, opened 2020-03-16
Adds new documentation for Alpha validator

froschdesign:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Enhancement        | yes


docs/book/validators/alpha.md
```diff
@@ -0,0 +1,80 @@
+# Alpha
+
+`Laminas\I18n\Validator\Alpha` allows you to validate if a given value
+**contains only alphabetical characters**. There is no length limitation for the
+input you want to validate. This validator is identical to the
+`Laminas\I18n\Validator\Alnum` validator with the exception that it does not
```
michalbundyra: should we link to the documentation of the Alnum validator here?

docs/book/validators/alpha.md
```diff
@@ -0,0 +1,80 @@
...
+## Basic Usage
+
+```php
+$validator = new Laminas\I18n\Validator\Alpha();
+
+if ($validator->isValid('Abcd')) {
+    // value contains only allowed chars
+} else {
+    // false
+}
```
michalbundyra: maybe better provide examples as in other validators documentation:
```php
var_dump($validator->isValid('Abcd')); // true
var_dump($validator->isValid('Virus24')); // false
```

docs/book/validators/alpha.md
```diff
@@ -0,0 +1,80 @@
...
+instantiation or afterwards via the `setAllowWhiteSpace()` method.
+
+```php fct_label="Constructor Usage"
+$validator = new Laminas\I18n\Validator\Alpha(['allowWhiteSpace' => true]);
+
+if ($validator->isValid('Abcd and efg')) {
+    // value contains only allowed chars
+} else {
+    // false
+}
```
michalbundyra: The same about examples as above

docs/book/validators/alpha.md
```diff
@@ -0,0 +1,80 @@
...
+
+The default value of this option is `false` that means whitespace characters are
+not allowed.
+
+## Using different Languages
+
+When using `Laminas\I18n\Validator\Alpha`, the language provided by the user's
+browser will be used to set the allowed characters. For locales outside of
+English, this means that additional alphabetic characters may be used
+&mdash; such as `ä`, `ö` and `ü` from the German alphabet.
```
michalbundyra: should we use normal `-` here?

docs/book/validators/alpha.md
froschdesign: I'm still not sure if I like the `var_dump` here. Other options are:

```php
$validator->isValid('Abcd'); // true
```
```php
$result = $validator->isValid('Abcd'); // true
```
```php
if ($validator->isValid('Abcd')) {
    // value contains only allowed chars
}
```
…

docs/book/validators/alpha.md
froschdesign: I think we should remove the dash for simplification and the different meanings / handlings in languages. See also: "[En dash versus em dash](https://en.wikipedia.org/wiki/Dash#En_dash_versus_em_dash)"

docs/book/validators/alpha.md
michalbundyra: ok, maybe if statement is the best then, but I guess we should remove `else { ...`

froschdesign: > maybe if statement is the best

I quite agree.

michalbundyra:
Thanks, @froschdesign!

PR #54, opened 2021-08-13
Avoid false return from substr() in  PHP < 8.0

### LINKED ISSUE #52: Issue with strpos call in laminas-stdlib when using IsFloat validator
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | latest

#### Summary

Initial misreported bug https://github.com/laminas/laminas-stdlib/issues/27

It seems when using this library in conjunction with laminas/laminas-stdlib:3.4.0, there is the potential for a boolean argument to be passed into the underlying wrappers `strpos` method, causing an exception to be thrown.

#### Current behavior

\Laminas\I18n\Validator\IsFloat::isValid:243 calls `strpos` and passes in $lastStringGroup which has the potential to be a boolean, causing an exception.

https://github.com/laminas/laminas-i18n/blob/b85a640138687f26129ace789fd24bbe0e0972ad/src/Validator/IsFloat.php#L240-L246

#### How to reproduce

Call \Laminas\I18n\Validator\IsFloat::isValid and pass in a string value (IE "1")

#### Expected behavior

\Laminas\I18n\Validator\IsFloat::isValid should not throw an exception

arokettu:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

Alternative way to fix #52

### DIFF
--- a/src/Validator/IsFloat.php
+++ b/src/Validator/IsFloat.php
@@ -235,7 +235,9 @@ class IsFloat extends AbstractValidator
         // No strrpos() in wrappers yet. ICU 4.x doesn't have grouping size for
         // everything. ICU 52 has 3 for ALL locales.
         $groupSize = $formatter->getAttribute(NumberFormatter::GROUPING_SIZE) ?: 3;
-        $lastStringGroup = $this->wrapper->substr($value, -$groupSize);
+        $lastStringGroup = $this->wrapper->strlen($value) > $groupSize ?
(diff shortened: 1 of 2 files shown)


froschdesign (approved):

froschdesign:
@arokettu 
Thank you for your time and this contribution! 👍

PR #69, opened 2022-02-20
Drops PHP 7.3 and Checks for the Intl Extension

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | kindof
| BC Break      | no
| QA            | yes

### Description

Sorry - this is quite large - Initially, I intended to take a look and see what can be done to decouple this from view, but there's quite a lot of tidying up to do first.

Initially, I dropped 7.3 and updated the lock on 7.4 - I also noticed that `intl` is declared in composer so I went through the component and removed all the conditionals that check for the extension.

I discovered that tests were failing for me in unexpected ways that was due to the default locale getting changed and not correctly restored/preserved between tests, so ended up making an abstract parent test case for i18n that preserves and restores the default locale on setUp() and tearDown()

I've also set phpunit to convert deprecations to exceptions and added the prophecy trait to dev deps to squash the deprecations from PHPUnit 9

<del>Also, CS passes locally, but code sniffer itself is chucking out deprecations on 7.4, so we need to upgrade LCS to whatever is current. I can do that here if it suits?</del>

Updated LCS to 2.3 and fixed all the violations 😬

… Also … a bit naughty, but I updated the workflows for renovate bot and set platform to 7.4


composer.json
```diff
@@ -16,7 +16,13 @@
         "forum": "https://discourse.laminas.dev"
     },
     "config": {
-        "sort-packages": true
+        "sort-packages": true,
+        "platform": {
+            "php": "7.4"
```
Ocramius: This means `7.4.0`, and `7.4.0` had a bunch of bugs in it :D 

IMO `7.4.99` or such

src/Filter/Alpha.php
```diff
@@ -1,10 +1,17 @@
 <?php
 
+declare(strict_types=1);
```
Ocramius: Won't this crash on scalar integers, for example?

src/Translator/Loader/Gettext.php
```diff
@@ -136,7 +150,7 @@ public function load($locale, $filename)
             $rawHeaders = explode("\n", trim($textDomain['']));
 
             foreach ($rawHeaders as $rawHeader) {
-                list($header, $content) = explode(':', $rawHeader, 2);
+                [$header, $content] = explode(':', $rawHeader, 2);
```
Ocramius, line 151: Is `$content` guaranteed to exist?

src/Translator/Loader/PhpMemoryArray.php
```diff
@@ -1,21 +1,26 @@
...
  */
 class PhpMemoryArray implements RemoteLoaderInterface
 {
-    /**
-     * @var array
-     */
+    /** @var array */
     protected $messages;
 
+    /** @param array $messages */
```
Ocramius, line 21: Could we be more precise about the type of `$messages`, perhaps?

src/Translator/Plural/Parser.php
```diff
@@ -219,7 +224,7 @@ protected function registerSymbol($id, $leftBindingPower = 0)
      */
     protected function getSymbol($id)
     {
-        if (! isset($this->symbolTable[$id])) {
+        if (! isset($this->symbolTable[$id])) { // phpcs:ignore
```
Ocramius, line 225: Wait what?

src/Translator/Plural/Parser.php
```diff
@@ -320,7 +325,7 @@ protected function getNextToken()
                 if ($this->string[$this->currentPos] === $result) {
                     $this->currentPos++;
                     $id = $result . $result;
-                } else {
+                } else { // phpcs:ignore
```
Ocramius, line 326: What's going on here? :|

src/Validator/Alnum.php
```diff
@@ -100,7 +108,7 @@ public function isValid($value)
 
         static::$filter->setAllowWhiteSpace($this->options['allowWhiteSpace']);
 
-        if ($value != static::$filter->filter($value)) {
+        if ($value != static::$filter->filter($value)) { // phpcs:ignore
```
Ocramius, line 109: Not sure why this needs to be ignored

src/View/Helper/DateFormat.php
```diff
@@ -123,7 +116,7 @@ public function setTimezone($timezone)
         $this->timezone = (string) $timezone;
 
         // The method setTimeZoneId is deprecated as of PHP 5.5.0
-        $setTimeZoneMethodName = (PHP_VERSION_ID < 50500) ? 'setTimeZoneId' : 'setTimeZone';
+        $setTimeZoneMethodName = PHP_VERSION_ID < 50500 ? 'setTimeZoneId' : 'setTimeZone';
```
Ocramius: This `PHP_VERSION_ID` check can be dropped

Ocramius (changes requested):
I went half-way with the review: the diff is too risky, in my opinion.

We would first need #37 to be part of this repo, and then we would need to hunt for clear type violations, before merging this.

As it currently is, the addition of `declare(strict_types=1)` on this codebase is too much.

It is also possible to just strip out that one CS rule for now: that would make the change much safer to push through.

src/Validator/Alnum.php
gsteel: Non-strict comparison

src/Translator/Plural/Parser.php
gsteel: I didn't want to touch existing source in this pull if I could avoid it: This one is an empty else

src/Translator/Plural/Parser.php
gsteel: I didn't want to touch existing source in this pull if I could avoid it: This one is an `empty if`

gsteel:
> I went half-way with the review: the diff is too risky, in my opinion.
> 
> We would first need #37 to be part of this repo, and then we would need to hunt for clear type violations, before merging this.
> 
> As it currently is, the addition of `declare(strict_types=1)` on this codebase is too much.
> 
> It is also possible to just strip out that one CS rule for now: that would make the change much safer to push through.

Started looking at the review prior to seeing your comment… Understood. It was pretty cavalier to put so many changes into one pull…

I'll send in a pull for psalm to replace #37 as it looks like it might be dead and see what I can resurrect from here later on…

Ocramius:
TBH, if you send in this patch without `declare(strict_types=1)`, that's already quite OK IMO

gsteel:
@Ocramius - #71 just landed - if that's OK to merge, I'll rebase this on that, update the baseline and strip out the strict types then?

src/Translator/Loader/PhpMemoryArray.php
gsteel: Correctly annotating messages might be a job for another day - I believe it's a shape that is returned by all the "loaders" and has a very specific shape that would benefit from a type on an interface somewhere…

src/Translator/Loader/Gettext.php
gsteel: I believe that gettext headers are `key:maybe-value` pairs - shall I revert this line? tests still pass and that line _is_ covered

Ocramius: The reason why I commented on this is that `declare(strict_types=1)` leads to a potential `$content = null` being passed to a function later on in this method.

gsteel:
@Ocramius and @froschdesign - I've reverted strict_types in src/ and rebased on the psalm merge.

Psalm errors are way down but some of the data providers were making my eyes bleed so I've glossed over a bit on the baseline. I've also addressed the previous review comments where possible.

src/Translator/Loader/Gettext.php
Ocramius: No longer applicable, since `declare(strict_types=1)` is gone.

src/Validator/DateTime.php
```diff
@@ -312,7 +298,7 @@ protected function getIntlDateFormatter()
                 $this->getTimeType(),
                 $this->timezone,
                 $this->calendar,
-                $this->pattern
+                $this->pattern ?? ''
```
Ocramius, line 301: Is an empty pattern acceptable?

Ocramius (approved):

src/Validator/DateTime.php
Ocramius: @gsteel only this comment left, before merge :)

gsteel: I coalesced this to an empty string from null because strict_types caused errors. The existing tests already cover `null` but I've added a couple of extra tests to make sure that an empty string has the same behaviour as a null.

Ocramius (approved):
Awesome, thanks @gsteel!

PR #75, opened 2022-02-24
Disconnect view helper inheritance hierarchy

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | yes and no…
| QA            | yes

### Description

This pull re-implements `AbstractHelper` with `getView` and `setView` and deprecates it at the same time, marking it as internal.

Whilst behaviour is preserved, consumers should be testing for the `HelperInterface` but they are probably testing for `AbstractHelper` in the View namespace, therefore this is probably a BC Break.

An alternative approach would be to first upgrade to view `2.20.0` which isn't released yet. 2.20 comes with [a trait that deprecates `getView` and `setView`](https://github.com/gsteel/laminas-view/blob/2.20.x/src/Helper/DeprecatedAbstractHelperHierarchyTrait.php).

### DIFF
--- a/src/View/Helper/AbstractHelper.php
+++ b/src/View/Helper/AbstractHelper.php
@@ -1,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\I18n\View\Helper;
+
+use Laminas\View\Helper\HelperInterface;
+use Laminas\View\Renderer\RendererInterface;
+
+/**
+ * @deprecated since >= 2.15 If it is necessary to access the renderer from within the plugin, you should inject it
+ *             as a constructor dependency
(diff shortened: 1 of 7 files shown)


froschdesign:
> An alternative approach would be to first upgrade to view `2.20.0` which isn't released yet. 2.20 comes with [a trait that deprecates `getView` and `setView`](https://github.com/gsteel/laminas-view/blob/2.20.x/src/Helper/DeprecatedAbstractHelperHierarchyTrait.php).

This sounds like a good idea.

gsteel:
I'll send in another pull when 2.20 is out and then close this. I'll keep it hanging around for now if that's OK to serve as a reminder.

PR #76, opened 2022-03-28
Deprecate Abstract View Helper Inheritance

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Bumps the minimum version of laminas-view to 2.20 in dev and imports the hierarchy deprecation trait into existing view helpers.

Closes #75 

The main problem with this pull is that `laminas-view` was/is listed as a dev dependency so there is no way of requiring a minimum version is installed. This will cause errors if view < 2.20 is installed because the trait imported in the i18n view helpers will not exist.

I think the choice here is to either re-consider #75 _(Perhaps make the introduced abstract helper extend that from view itself)_, or list laminas-view as a hard dependency here at a 2.20 minimum, which sort of defeats the point - removing the inheritance chain enables i18n to be completely decoupled from view.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,8 @@
             "php": "7.4.99"
         },
         "allow-plugins": {
-            "dealerdirect/phpcodesniffer-composer-installer": true
+            "dealerdirect/phpcodesniffer-composer-installer": true,
+            "composer/package-versions-deprecated": true
         }
     },
     "extra": {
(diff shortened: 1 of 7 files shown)


Ocramius (approved):

Ocramius:
> This will cause errors if view < 2.20 is installed because the trait imported in the i18n view helpers will not exist.

We can add a `conflict:` clause

gsteel:
> We can add a `conflict:` clause

Oh yeah! Gimme a sec ;)

Ocramius (approved):

PR #78, opened 2022-07-12
General static analysis improvements and Removal of Interop\Container

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

Note that `laminas-servicemanager` is not a hard dependency here, but nothing has changed that might cause issues with v2. 🤞

### DIFF
--- a/src/Translator/LoaderPluginManager.php
+++ b/src/Translator/LoaderPluginManager.php
@@ -124,6 +124,9 @@ class LoaderPluginManager extends AbstractPluginManager
      *
      * Proxies to `validate()`.
      *
+     * @deprecated Since 2.16.0 - This component is no longer compatible with Service Manager v2.
+     *             This method will be removed in version 3.0
+     *
      * @param mixed $plugin
      * @throws Exception\RuntimeException
      * @psalm-assert InstanceType $plugin
--- a/src/Translator/LoaderPluginManager.php
+++ b/src/Translator/LoaderPluginManager.php
@@ -3,6 +3,8 @@
 namespace Laminas\I18n\Translator;
 
 use Laminas\I18n\Exception;
+use Laminas\I18n\Translator\Loader\FileLoaderInterface;
+use Laminas\I18n\Translator\Loader\RemoteLoaderInterface;
 use Laminas\ServiceManager\AbstractPluginManager;
 use Laminas\ServiceManager\Exception\InvalidServiceException;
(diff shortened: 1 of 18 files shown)


src/Translator/LoaderPluginManagerFactory.php
```diff
@@ -2,10 +2,10 @@
 
 namespace Laminas\I18n\Translator;
 
-use Interop\Container\ContainerInterface;
 use Laminas\ServiceManager\Config;
 use Laminas\ServiceManager\FactoryInterface;
 use Laminas\ServiceManager\ServiceLocatorInterface;
+use Psr\Container\ContainerInterface;
```
Ocramius, line 8: I think this requires a conflict with earlier ServiceManager versions?

gsteel: That's probably anything prior to 3.14 right?

src/Translator/LoaderPluginManager.php
```diff
@@ -100,20 +102,20 @@ class LoaderPluginManager extends AbstractPluginManager
...
             return;
         }
 
         throw new InvalidServiceException(sprintf(
-            'Plugin of type %s is invalid; must implement %s\Loader\FileLoaderInterface '
-            . 'or %s\Loader\RemoteLoaderInterface',
+            'Plugin of type %s is invalid; must implement %s or %s',
             is_object($plugin) ? get_class($plugin) : gettype($plugin),
-            __NAMESPACE__,
-            __NAMESPACE__
```
Ocramius, line 116: Interesting how these were completely wrong :O

src/Translator/LoaderPluginManagerFactory.php
Ocramius: Hmm, thinking that this change is a BC break **unless** we require the servicemanager component (which declares the interop aliases, therefore providing a BC layer)

gsteel: It makes sense to me to make it a hard dependency due to the plugin manager. Just say the word and I'll revert or add the dependency

Ocramius: I think adding the dependency makes more sense anyway

Ocramius (approved):

Ocramius:
Thanks @gsteel!

PR #84, opened 2022-08-08
Adds Country Code Value Object

gsteel:
|    Q          |   A
|-------------- | ------
| New Feature   | yes

### Description

As discussed [in laminas-i18n-phone-number#1](https://github.com/laminas/laminas-i18n-phone-number/pull/1#pullrequestreview-1064960141), this patch introduces a Country Code value object, initially for use with the forthcoming phone number validation library.

PHP is bumped to 8.1 minimum to take advantage of readonly public properties

### DIFF
--- a/src/CountryCode.php
+++ b/src/CountryCode.php
@@ -89,7 +89,7 @@ final class CountryCode
      * @throws InvalidArgumentException When a non-empty string is provided that cannot be recognized,
      *                                  an exception will be thrown.
      */
-    public static function detect(string|self|null $countryCodeOrLocale): self
+    public static function detect(string|self|null $countryCodeOrLocale = null): self
     {
         if ($countryCodeOrLocale instanceof self) {
             return $countryCodeOrLocale;
--- a/src/CountryCode.php
+++ b/src/CountryCode.php
@@ -1,0 +1,137 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\I18n;
+
+use Laminas\I18n\Exception\InvalidArgumentException;
+use Locale;
+
+use function assert;
+use function preg_match;
+use function strtoupper;
+
+/**
+ * @psalm-immutable
+ */
+final class CountryCode
+{
+    /** @param non-empty-string $code */
+    private function __construct(public readonly string $code)
+    {
+    }
+
+    /** @return non-empty-string */
+    public function toString(): string
+    {
+        return $this->code;
+    }
+
+    public function equals(self $other): bool
+    {
+        return $this->code === $other->code;
+    }
+
+    /**
+     * Create a new ValueObject from an ISO 3166 Country Code
+     * Country codes are 2 letter, uppercase strings representing a country identifier on planet earth. The given
+     * value must also represent a country known by PHP’s intl extension.
+     * Valid values include 'US', 'GB', 'ZA', 'FR' etc.
+     *
+     * @link https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes
+     *
+     * @param non-empty-string $code
+     * @throws InvalidArgumentException An invalid string or an unknown country will cause an exception.
+     * @psalm-pure
+     */
+    public static function fromString(string $code): self
+    {
+        $code = strtoupper($code);
+        if (! preg_match('/^[A-Z]{2}$/', $code)) {
+            throw InvalidArgumentException::withInvalidCountryCode($code);
+        }
+
+        $displayName = Locale::getDisplayRegion('-' . $code, 'GB');
+        if ($displayName === '' || $displayName === 'Unknown Region') {
+            throw InvalidArgumentException::withUnknownCountryCode($code);
+        }
+
+        return new self($code);
+    }
+
+    /**
+     * Create a new value object from a locale string
+     *
+     * Given a well-formed locale, this method will extract the relevant country code and proxy to @link fromString
+     * Valid values include: 'en_GB', 'en-GB', 'zh-Hans-CN'
+     *
+     * @param non-empty-string $locale
+     * @throws InvalidArgumentException An unrecognizable locale will cause an exception.
+     * @psalm-pure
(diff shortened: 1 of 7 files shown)


froschdesign:
@gsteel 
> PHP is bumped to 8.1 minimum to take advantage of readonly public properties

I suggest using 8.0, because limiting the whole component to the latest version for this _one_ value object, which is not actively used here at the moment, is too much.
Version 8.1 for a new component like the related laminas-i18n-phone-number is the right way but not for this existing component without other changes.

internalsystemerror (approved):
LGTM

froschdesign:
> @gsteel
> 
> > PHP is bumped to 8.1 minimum to take advantage of readonly public properties
> 
> I suggest using 8.0, because limiting the whole component to the latest version for this _one_ value object, which is not actively used here at the moment, is too much. Version 8.1 for a new component like the related laminas-i18n-phone-number is the right way but not for this existing component without other changes.

internalsystemerror:
Addressing:

> I suggest using 8.0, because limiting the whole component to the latest version for this one value object, which is not actively used here at the moment, is too much.

Maybe mark the VO as internal for now, using private properties in the constructor and simple getter methods? Then when 8.1 support is added at a later date, we can make the properties readonly public again and remove the internal tag?

froschdesign:
> Maybe mark the VO as internal for now…

Internal would not work here because it should be used by other components and the user.

> …using private properties in the constructor and simple getter methods?

👍🏻

gsteel:
OK! 😢 - I just really like `public readonly`. Rather than a getter here, I think it should implement `Stringable` and `__toString` though.

internalsystemerror:
> I just really like public readonly.

Same, I'm quite eager to be able to use this, esp for things like VO's where cloning isn't necessary.

> Internal would not work here because it should be used by other components and the user.

Is there another option that would let us make the change to public readonly properties later without requiring a new major?

froschdesign:
> OK! 😢 - I just really like `public readonly`.

I like it too, but with the current state of the component, it's not very helpful to discard version 8 completely.

gsteel:
> > Internal would not work here because it should be used by other components and the user.
> 
> Is there another option that would let us make the change to public readonly properties later without requiring a new major?

I've just made it private for now. Changing the private prop to `public readonly` at a later date wouldn't break BC. Also, there's no need for `Stringable` or another getter because it already has a `toString` public method

internalsystemerror:
> Changing the private prop to public readonly at a later date wouldn't break BC

No but changing usages from `$obj->code()` to `$obj->code` would be, but I guess we could keep the getters anyway which I didn't consider.

src/CountryCode.php
```diff
@@ -17,7 +17,7 @@
 final class CountryCode
 {
     /** @param non-empty-string $code */
-    private function __construct(public readonly string $code)
+    private function __construct(private string $code)
```
internalsystemerror, line 20: Since this VO has a single property, we can get away without a specific getter. I'm thinking about consistency with a possible additional future VO with 2 properties say.

Solve that when we come to it? Hopefully we can rely on `public readonly` by then?

gsteel: In this case, the value is a single string - it makes it pretty easy and intuitive to go back and forth between a string and object, i.e `CountryCode::fromString($code->toString())`.

If this VO evolves at all, I could see it possibly attracting methods such as `$countryCode->getDisplayName('en-GB'); // "United Kingdom"`, but I can't envision additional props right now.

internalsystemerror: I meant if we add another VO for something else that has multiple properties. Then you would expect to get the various properties of the new VO via the getters (which would be inconsistent with this VO), unless we are using public properties by that point.

Don't consider this a blocker, my approval stands. I'm worrying about a problem that isn't a problem yet.

gsteel:
@Ocramius & @froschdesign - Bearing in mind that this patch drops 7.4 support, could I have a review on this one please if anyone has the time?

src/CountryCode.php
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Return a country code from either a string code or a locale string falling back to the system locale if null
+     *
+     * @link fromLocaleString
+     * @link fromString
+     *
+     * @throws InvalidArgumentException When a non-empty string is provided that cannot be recognized,
+     *                                  an exception will be thrown.
+     */
+    public static function detect(string|self|null $countryCodeOrLocale): self
```
Ocramius: Should this be `= null`, so that `CountryCode::detect()` is a more explicit call to the system locale logic? :thinking:

test/CountryCodeTest.php
```diff
@@ -0,0 +1,145 @@
...
+declare(strict_types=1);
+
+namespace LaminasTest\I18n;
+
+use Laminas\I18n\CountryCode;
+use Laminas\I18n\Exception\InvalidArgumentException;
+use Locale;
+use PHPUnit\Framework\TestCase;
+
+final class CountryCodeTest extends TestCase
```
Ocramius, line 13: Let's add an `@covers` here

Ocramius (approved):
LGTM - also with dropping PHP version :+1:

Ocramius:
Ah, one thing is missing here: where would this be used? Or is it "just sitting there"? :D 

A short docs chapter with high-level use-case would help a lot here.

gsteel:
Thanks for reviewing @ocramius - The value object is currently, specifically for i18n-phone-number - @froschdesign asked to move it here - that said, I have a pull in the works that makes use of it, assuming it gets merged. It would be easier to update docs with that if OK? I'll update covers annotations in a sec.

Ocramius:
Works for me :+1:

src/CountryCode.php
gsteel: Yes - I think a default null would be helpful here 👍

Ocramius:
Feel free to release whenever ready then :)

PR #87, opened 2022-10-10
Add support for PHP 8.2

gsteel:
|    Q          |   A
|-------------- | ------
| New Feature   | yes

### Description

Updates composer.json to support 8.2 - All direct deps already support 8.2, but dev deps do not, so I've also ignored platform reqs here


Ocramius (approved):
LGTM :+1:

PR #86, opened 2022-09-13
Add `CountryCodeDataList` view helper and `CountryCode` validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | not yet - required
| New Feature   | yes
| RFC           | no?

### Description

Given the addition of the country code value object in #84, this patch puts it to use by adding a list object available in the container, that users can use to retrieve a list of all ISO-3166 alpha 2 country codes. The list implements and is aliased to a simple interface that user can override in the container, for example, in order to restrict the list to desired countries.

A view helper is added that outputs an HTML `datalist` element, useful for autocomplete and a validator for ISO3166 country codes.

Documentation is outstanding.

PR #89, opened 2022-12-02
Bump dev dependencies and improve config related type inference

gsteel:
This should help get #86 over the line


Ocramius (approved):
Thanks @gsteel

PR #109, opened 2023-11-03
Adds PHP 8.3 as new supported version

fezfez:
replace https://github.com/laminas/laminas-i18n/pull/108


gsteel (approved):
Thank you @fezfez

PR #112, opened 2023-11-13
Minimum number of decimals in View-Helper NumberFormat

MatthiasKuehneEllerhold:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | yes
| QA            | no

### Description

We have some user provided numbers that may have zero to endless decimals. We want to display them in the user locale exactly as provided.
If the user inputs "1", we want to display "1". If the user inputs "0,5432" (german notation!) we want to display "0,5432".
Sadly with the current (2.24.1) View-Helper we have to know beforehand how many decimals the user has used.

The underlying NumberFormatter-Class can do this: by setting min-decimals to zero and max-decimals to PHP_INT_MAX (or some large arbitrary number). The view-helper does not allow to set these 2 options separatly (yet). Thats what I added.

According to the github-template I should target the "develop" branch. Sadly no branch named "develop" can be found in this repository, so I've targeted the branch of the currently active minor version. Please advice if this was the right approach.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -938,16 +938,16 @@
         },
         {
             "name": "laminas/laminas-cache",
-            "version": "3.11.0",
+            "version": "3.12.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/laminas/laminas-cache.git",
-                "reference": "76cc49af422a9cf5deff9cd0f63cbda29ae88d14"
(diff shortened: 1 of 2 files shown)


Ocramius:
@MatthiasKuehneEllerhold could you perhaps express the problem in a failing test first?

PR #113, opened 2023-11-13
Add the possibility to set the minimum number of decimals in the numb…

MatthiasKuehneEllerhold:
Q | A
-- | --
Documentation | no
Bugfix | no
BC Break | no
New Feature | yes
RFC | yes
QA | no

## Description
We have some user provided numbers that may have zero to 
endless decimals. We want to display them in the user locale exactly as 
provided.

If the user inputs "1", we want to display "1". If the user inputs "0,5432" (german notation!) we want to display "0,5432".

Sadly with the current (2.24.1) View-Helper we have to know beforehand how many decimals the user has used.</p>

The underlying NumberFormatter-Class can do this: by 
setting min-decimals to zero and max-decimals to PHP_INT_MAX (or some 
large arbitrary number). The view-helper does not allow to set these 2 
options separatly (yet). Thats what I added.

According to the github-template I should target the 
"develop" branch. Sadly no branch named "develop" can be found in this 
repository, so I've targeted the branch of the currently active minor 
version. Please advice if this was the right approach.

Some commits are optional: adding type hint to variables and classes and may be discard if undesired.

### DIFF
--- a/src/View/Helper/NumberFormat.php
+++ b/src/View/Helper/NumberFormat.php
@@ -60,8 +60,6 @@ class NumberFormat extends AbstractHelper
      * Format a number
      *
      * @param  int|float   $number
-     *
-     * @return string
      */
     public function __invoke(
         $number,
@@ -71,7 +69,7 @@ class NumberFormat extends AbstractHelper
         ?int $maxDecimals = null,
         ?array $textAttributes = null,
         ?int $minDecimals = null
-    ) {
+    ): string {
         if (null === $locale) {
             $locale = $this->getLocale();
         }
@@ -125,9 +123,6 @@ class NumberFormat extends AbstractHelper
 
     /**
      * Set format style to use instead of the default
-     *
-     * @param  int $formatStyle
-     * @return $this
      */
     public function setFormatStyle(int $formatStyle): self
     {
@@ -137,8 +132,6 @@ class NumberFormat extends AbstractHelper
 
     /**
      * Get the format style to use
-     *
-     * @return int
      */
     public function getFormatStyle(): int
     {
@@ -151,9 +144,6 @@ class NumberFormat extends AbstractHelper
 
     /**
      * Set format type to use instead of the default
-     *
-     * @param  int $formatType
-     * @return $this
      */
     public function setFormatType($formatType): self
     {
@@ -163,8 +153,6 @@ class NumberFormat extends AbstractHelper
 
     /**
      * Get the format type to use
-     *
-     * @return int
      */
     public function getFormatType(): int
     {
@@ -176,9 +164,6 @@ class NumberFormat extends AbstractHelper
 
     /**
      * Set number (both min & max) of decimals to use instead of the default.
-     *
-     * @param  int|null $decimals
-     * @return $this
      */
     public function setDecimals(?int $decimals): self
     {
@@ -189,9 +174,6 @@ class NumberFormat extends AbstractHelper
 
     /**
      * Set the maximum number of decimals to use instead of the default.
-     *
-     * @param  int|null $maxDecimals
-     * @return $this
      */
     public function setMaxDecimals(?int $maxDecimals): self
     {
@@ -201,9 +183,6 @@ class NumberFormat extends AbstractHelper
 
     /**
      * Set the minimum number of decimals to use instead of the default.
-     *
-     * @param int|null $minDecimals
-     * @return $this
      */
     public function setMinDecimals(?int $minDecimals): self
     {
@@ -213,8 +192,6 @@ class NumberFormat extends AbstractHelper
 
     /**
      * Get number of decimals.
-     *
-     * @return int|null
      */
     public function getDecimals(): ?int
     {
@@ -223,8 +200,6 @@ class NumberFormat extends AbstractHelper
 
     /**
      * Get the maximum number of decimals.
-     *
-     * @return int|null
      */
     public function getMaxDecimals(): ?int
     {
@@ -233,8 +208,6 @@ class NumberFormat extends AbstractHelper
 
     /**
      * Get the minimum number of decimals.
-     *
-     * @return int|null
      */
     public function getMinDecimals(): ?int
     {
@@ -243,20 +216,15 @@ class NumberFormat extends AbstractHelper
 
     /**
      * Set locale to use instead of the default.
-     *
-     * @param  string $locale
-     * @return $this
      */
-    public function setLocale($locale): self
+    public function setLocale(?string $locale): self
     {
-        $this->locale = (string) $locale;
+        $this->locale = $locale;
         return $this;
     }
 
     /**
      * Get the locale to use
-     *
-     * @return string
      */
     public function getLocale(): string
     {
@@ -267,18 +235,11 @@ class NumberFormat extends AbstractHelper
         return $this->locale;
     }
 
-    /**
-     * @return array
-     */
     public function getTextAttributes(): array
     {
         return $this->textAttributes;
     }
 
-    /**
-     * @param array $textAttributes
-     * @return $this
-     */
     public function setTextAttributes(array $textAttributes): self
     {
         $this->textAttributes = $textAttributes;
--- a/src/View/Helper/NumberFormat.php
+++ b/src/View/Helper/NumberFormat.php
@@ -20,11 +20,18 @@ class NumberFormat extends AbstractHelper
     use DeprecatedAbstractHelperHierarchyTrait;
 
     /**
-     * number of decimals to use.
+     * The maximum number of decimals to use.
      *
      * @var int
      */
-    protected $decimals;
+    protected $maxDecimals;
+
+    /**
+     * The minimum number of decimals to use.
+     *
+     * @var int
+     */
+    protected $minDecimals;
 
     /**
      * NumberFormat style to use
@@ -68,8 +75,10 @@ class NumberFormat extends AbstractHelper
      * @param  int|null    $formatStyle
      * @param  int|null    $formatType
      * @param  string|null $locale
-     * @param  int|null    $decimals
+     * @param  int|null    $maxDecimals
      * @param  array|null  $textAttributes
+     * @param  int|null     $minDecimals
+     *
      * @return string
      */
     public function __invoke(
@@ -77,8 +86,9 @@ class NumberFormat extends AbstractHelper
         $formatStyle = null,
         $formatType = null,
         $locale = null,
-        $decimals = null,
-        ?array $textAttributes = null
+        $maxDecimals = null,
+        ?array $textAttributes = null,
+        ?int $minDecimals = null
     ) {
         if (null === $locale) {
             $locale = $this->getLocale();
@@ -89,15 +99,22 @@ class NumberFormat extends AbstractHelper
         if (null === $formatType) {
             $formatType = $this->getFormatType();
         }
-        if (! is_int($decimals) || $decimals < 0) {
-            $decimals = $this->getDecimals();
+        if (! is_int($minDecimals) || $minDecimals < 0) {
+            $minDecimals = $this->getMinDecimals();
+        }
+        if (! is_int($maxDecimals) || $maxDecimals < 0) {
+            $maxDecimals = $this->getMaxDecimals();
+        }
+        if (($maxDecimals !== null) && ($minDecimals === null)) {
+            // Fallback to old behavior
+            $minDecimals = $maxDecimals;
         }
         if (! is_array($textAttributes)) {
             $textAttributes = $this->getTextAttributes();
         }
 
         $formatterId = md5(
-            $formatStyle . "\0" . $locale . "\0" . $decimals . "\0"
+            $formatStyle . "\0" . $locale . "\0" . $minDecimals . "\0" . $maxDecimals . "\0"
             . md5(serialize($textAttributes))
         );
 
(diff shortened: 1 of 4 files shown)


src/View/Helper/NumberFormat.php
```diff
@@ -13,73 +15,66 @@
 use function serialize;
 
 /**
- * View helper for formatting dates.
+ * View helper for formatting numbers.
```
Ocramius, line 16: Let's first declare a test scenario where the issue is visible, then adjust the code afterwards :-)

Ocramius (changes requested):

src/View/Helper/NumberFormat.php
MatthiasKuehneEllerhold: For this change in particular? ;-)
I'll add a test for the min/max decimals shortly. :)

MatthiasKuehneEllerhold:
Added Unit-Tests and fixed CI (hopefully). DCO still says "signoff missing" although all commits are signed by my GPG key. What did I do wrong?
Additionally: Im not sure on how to fix the remaining psalm errors. Any hint please?

src/View/Helper/NumberFormat.php
```diff
@@ -13,73 +15,66 @@
...
 {
     use DeprecatedAbstractHelperHierarchyTrait;
 
     /**
-     * number of decimals to use.
-     *
-     * @var int
+     * The maximum number of decimals to use.
+     */
+    protected ?int $maxDecimals = null;
```
gsteel: This newly added property can be private - ideally, all of the properties in this class would be private, but it is not possible to make them so without breaking BC - this is because the class has not been marked as final and is therefore open to inheritance.

src/View/Helper/NumberFormat.php
```diff
@@ -13,73 +15,66 @@
...
+
+    /**
+     * The minimum number of decimals to use.
      */
-    protected $decimals;
+    protected ?int $minDecimals = null;
```
gsteel: This is a BC break for inheritors. Both the property type hint and the change to the property name will break any implementations that extend from this class.

The null default value is acceptable, but a docblock should be added such as `/** @var int|null **/`

src/View/Helper/NumberFormat.php
```diff
@@ -13,73 +15,66 @@
...
 
     /**
      * NumberFormat style to use
-     *
-     * @var int
      */
-    protected $formatStyle;
+    protected ?int $formatStyle = null;
```
gsteel: Property type addition here is a BC break

src/View/Helper/NumberFormat.php
```diff
@@ -13,73 +15,66 @@
...
 
     /**
      * NumberFormat type to use
-     *
-     * @var int
      */
-    protected $formatType;
+    protected ?int $formatType = null;
```
gsteel: Property type addition here is a BC break

src/View/Helper/NumberFormat.php
```diff
@@ -13,73 +15,66 @@
...
 
     /**
      * Formatter instances
      *
      * @var array<string, NumberFormatter>
      */
-    protected $formatters = [];
+    protected array $formatters = [];
```
gsteel: Property type addition here is a BC break

src/View/Helper/NumberFormat.php
```diff
@@ -13,73 +15,66 @@
...
 
     /**
      * Text attributes.
      *
-     * @var array
+     * @var array<int, string>
      */
-    protected $textAttributes = [];
+    protected array $textAttributes = [];
```
gsteel: Property type addition here is a BC break

src/View/Helper/NumberFormat.php
```diff
@@ -13,73 +15,66 @@
...
 
     /**
      * Locale to use instead of the default
-     *
-     * @var string
      */
-    protected $locale;
+    protected ?string $locale = null;
```
gsteel: Property type addition here is a BC break

src/View/Helper/NumberFormat.php
```diff
@@ -13,73 +15,66 @@
...
-        $decimals = null,
-        ?array $textAttributes = null
-    ) {
+        ?int $formatStyle = null,
+        ?int $formatType = null,
+        ?string $locale = null,
+        ?int $maxDecimals = null,
+        ?array $textAttributes = null,
+        ?int $minDecimals = null
+    ): string {
```
gsteel: Parameter type additions here are unfortunately BC breaks. Really, we should also avoid parameter name changes for BC, i.e. `$decimals` rather than `$maxDecimals` because of named parameters.

Removing the native types will necessitate re-adding the doc-block types, but these can be improved with more precise types such as `positive-int|0|null` for example - this will cause a raft of new psalm issues so you can just reinstate the existing docs assuming they are correct.

src/View/Helper/NumberFormat.php
```diff
@@ -123,22 +128,17 @@ public function __invoke(
 
     /**
      * Set format style to use instead of the default
-     *
-     * @param  int $formatStyle
-     * @return $this
      */
-    public function setFormatStyle($formatStyle)
+    public function setFormatStyle(int $formatStyle): self
```
gsteel: We can't add parameter or return types without breaking BC

src/View/Helper/NumberFormat.php
```diff
@@ -123,22 +128,17 @@ public function __invoke(
...
         return $this;
     }
 
     /**
      * Get the format style to use
-     *
-     * @return int
      */
-    public function getFormatStyle()
+    public function getFormatStyle(): int
```
gsteel: We can't add parameter or return types without breaking BC

src/View/Helper/NumberFormat.php
```diff
@@ -149,22 +149,17 @@ public function getFormatStyle()
 
     /**
      * Set format type to use instead of the default
-     *
-     * @param  int $formatType
-     * @return $this
      */
-    public function setFormatType($formatType)
+    public function setFormatType(?int $formatType): self
```
gsteel: We can't add parameter or return types without breaking BC

src/View/Helper/NumberFormat.php
```diff
@@ -149,22 +149,17 @@ public function getFormatStyle()
...
         return $this;
     }
 
     /**
      * Get the format type to use
-     *
-     * @return int
      */
-    public function getFormatType()
+    public function getFormatType(): int
```
gsteel: We can't add parameter or return types without breaking BC

src/View/Helper/NumberFormat.php
```diff
@@ -173,45 +168,70 @@ public function getFormatType()
...
         return $this;
     }
 
     /**
      * Get number of decimals.
-     *
-     * @return int
      */
-    public function getDecimals()
+    public function getDecimals(): ?int
```
gsteel: We can't add parameter or return types without breaking BC

src/View/Helper/NumberFormat.php
```diff
@@ -173,45 +168,70 @@ public function getFormatType()
...
     }
 
     /**
      * Set locale to use instead of the default.
-     *
-     * @param  string $locale
-     * @return $this
      */
-    public function setLocale($locale)
+    public function setLocale(?string $locale): self
```
gsteel: We can't add parameter or return types without breaking BC

src/View/Helper/NumberFormat.php
```diff
@@ -173,45 +168,70 @@ public function getFormatType()
...
         return $this;
     }
 
     /**
      * Get the locale to use
-     *
-     * @return string
      */
-    public function getLocale()
+    public function getLocale(): string
```
gsteel: We can't add parameter or return types without breaking BC

src/View/Helper/NumberFormat.php
```diff
@@ -221,18 +241,17 @@ public function getLocale()
     }
 
     /**
-     * @return array
+     * @return array<int, string>
      */
-    public function getTextAttributes()
+    public function getTextAttributes(): array
```
gsteel: We can't add parameter or return types without breaking BC

src/View/Helper/NumberFormat.php
```diff
@@ -221,18 +241,17 @@ public function getLocale()
...
         return $this->textAttributes;
     }
 
     /**
-     * @param array $textAttributes
-     * @return $this
+     * @param array<int, string> $textAttributes
      */
-    public function setTextAttributes(array $textAttributes)
+    public function setTextAttributes(array $textAttributes): self
```
gsteel: We can't add the return type here without breaking BC

test/View/Helper/NumberFormatTest.php
```diff
@@ -22,8 +22,8 @@ protected function setUp(): void
         $this->helper = new NumberFormatHelper();
     }
 
-    /** @return array<array-key, array{0: string, 1: int, 2: int, 3: int|null, 4: array<int, string>, 5: float, 6: string}> */
-    public static function currencyTestsDataProvider(): array
+    /** @return array<array-key, array{0: string, 1: int, 2: int, 3: int|null, 4: array<int, string>, 5: int|null, 6: float, 7: string}> */
+    public static function numberTestsDataProvider(): array
```
gsteel, line 26: Psalm will consider PHPUnit data providers as `PossiblyUnusedMethod` - the best thing to do here is add a suppression to `psalm.xml` for the whole `/test` directory.

Adding something like:
```xml
<issueHandlers>
  <PossiblyUnusedMethod>
    <errorLevel type="suppress">
      <directory name="test" />
    </errorLevel>
  </PossiblyUnusedMethod>
</issueHandlers>
```

The name change is a valid improvement BTW, it's just that, because of the name change, the error that is currently suppressed in the psalm baseline has re-surfaced.

src/View/Helper/NumberFormat.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
```
gsteel: This lib is probably not ready for `strict_types` yet. More context in https://github.com/laminas/laminas-i18n/pull/69

All of the `UnusedBaselineEntry` errors from psalm can be resolved with `vendor/bin/psalm --update-baseline`

src/View/Helper/NumberFormat.php
```diff
@@ -173,45 +168,70 @@ public function getFormatType()
...
+        return $this->maxDecimals;
+    }
+
+    /**
+     * Get the minimum number of decimals.
+     */
+    public function getMinDecimals(): ?int
+    {
+        return $this->minDecimals;
     }
```
gsteel: Are these getters necessary? It's better not to add new public methods when the _effect_ of the new `minDecimals` property can be tested without verifying the value of the internal property.

gsteel (changes requested):
Thanks @MatthiasKuehneEllerhold - Generally speaking, whilst a desirable improvement, adding the native property/return/parameter types to existing props and methods all break BC. Because we're targeting a minor, they will have to be reverted and doc-blocks reinstated. Sorry!

src/View/Helper/NumberFormat.php
MatthiasKuehneEllerhold: Ive added them as "protected" to keep consistent with the other attributes.
What I like about the zend framework then and laminas now is that I can overwrite and change classes and change behavior in subtle ways. Ideally I just overwrite a small piece of a class instead of copy and paste everything in a new class.
Thats why I'm personally in favour of making everything protected (and therefore accessable to inherited classes) instead of privating everything and forcing myself to copy-and-pasting whole classes.

src/View/Helper/NumberFormat.php
MatthiasKuehneEllerhold: Reverted.

src/View/Helper/NumberFormat.php
MatthiasKuehneEllerhold: Reverted

src/View/Helper/NumberFormat.php
MatthiasKuehneEllerhold: Reverted

src/View/Helper/NumberFormat.php
MatthiasKuehneEllerhold: Reverted

src/View/Helper/NumberFormat.php
MatthiasKuehneEllerhold: Reverted

src/View/Helper/NumberFormat.php
MatthiasKuehneEllerhold: Everything else had a getter, so I added them.
Are they necessary though? I have a few situations where I get all internal state from an object, change it and then set the old state back. e. g. while rendering a view script I get the old locale from a lot of entities, set the desired locale, render my PHTML and then re-set the old locale. 
Im not sure this situation would ever arise here though. But I erred on the side of caution and added them.

src/View/Helper/NumberFormat.php
MatthiasKuehneEllerhold: Reverted

src/View/Helper/NumberFormat.php
MatthiasKuehneEllerhold: Ive looked at the errors that psalm gives... while I'm tempted to fix em, this is not the right PR for it (as you've said so yourself.).
declare reverted.

MatthiasKuehneEllerhold:
I could add another PR with all the type hint changes, this time targeting a more recent branch - if you like. All it would take is a git revert of ae1da38 (and possibly some more work in the other files ;-) ).

PR #122, opened 2024-05-18
Expand documentation in introduction and validators files

### LINKED ISSUE #83: Document the use of intl extension
We may want to amend the tutorial and/or introduction page in the documentation to explicitly state the use of the `intl` extension, with links to the documentation on php.net.

A user may infer that we are abstracting a lot of this functionality when it can be used directly.

https://www.php.net/manual/en/book.intl.php

froschdesign:
|    Q          |   A
|-------------- | ------
| Documentation | yes

Fixes #83

### DIFF
--- a/docs/book/introduction.md
+++ b/docs/book/introduction.md
@@ -1,10 +1,35 @@
 # Introduction
 
-`Laminas\I18n` comes with a complete translation suite which supports all major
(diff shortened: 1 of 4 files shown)


gsteel (approved):
LGTM!

PR #123, opened 2024-05-18
Update caching documentation for laminas-cache

### LINKED ISSUE #10: Using a Filesystem cache storage adapter results in an exception
> `Object of class Zend\I18n\Translator\TextDomain could not be converted to string`

Switching to another adapter (like `memory`) works fine, looks like this happens because `Filesystem` doesn’t do any serialization, just:

``` php
// Zend\Cache\Storage\Adapter\Filesystem::putFileContent()
if (! is_string($data)) {
    // Ensure we have a string
    $data = (string) $data;
}
```

I’m not sure if `Translator` should serialize the `TextDomain` object before storing it, or if this is more a bug in the cache adapter. Perhaps we should check whether the cache can indeed store objects by using `$cache->getCapabilities()`, and serialize it if not?

---

Originally posted by @kinglozzer at https://github.com/zendframework/zend-i18n/issues/44

froschdesign:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | yes

Fixes #10

In the example for caching, the file system adapter is replaced by the memory adapter, as the file system adapter requires a serializer, which requires a further installation step via Composer. There are already some installation instructions and explanations.

### DIFF
--- a/docs/book/translator/caching.md
+++ b/docs/book/translator/caching.md
@@ -16,6 +16,15 @@ guarantees an optimized loading procedure.
 > ```bash
 > $ composer require laminas/laminas-cache-storage-deprecated-factory
 > ```
+>
+> laminas-cache is shipped without a specific cache adapter to allow free choice of storage backends and their dependencies.
+> So make sure that the required adapters are installed.
+>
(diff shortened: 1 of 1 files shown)


gsteel (approved):
Thanks @froschdesign

PR #125, opened 2024-06-17
Relax Service Manager Constraint

gsteel:
This patch allows installation of SM v4 - this lib is not compatible with SMv4 in any way yet.


composer.json
```diff
@@ -34,7 +34,7 @@
     "require": {
         "php": "~8.1.0 || ~8.2.0 || ~8.3.0",
         "ext-intl": "*",
-        "laminas/laminas-servicemanager": "^3.21.0",
+        "laminas/laminas-servicemanager": "^3.21.0 | ^4.0",
```
Ocramius: does CI actually verify `^4`, or does it just fail?

Ocramius: BTW, `||` (I think single `|` got deprecated ages ago)

gsteel: Thanks - I'm just getting a branch going for SMv4 because of cyclic dependencies nightmares…

gsteel:
With the release of Laminas\Translator https://github.com/laminas/laminas-translator/releases/tag/1.0.0 this patch is no longer required

PR #128, opened 2024-07-01
`Laminas\Translator`

gsteel:
|    Q          |   A
|-------------- | ------
| New Feature   | yes

### Description

Requires `laminas/laminas-translator` and extends the shipped interface via `Laminas\I18n\Translator\TranslatorInterface`

As a straight copy of the existing interface, there's no BC break, but I guess there is a "technical" BC break because of any change in inheritance.

If this *has* to go into a major release, it defeats the point of extracting the interface in the first place.

### DIFF
--- a/src/Translator/TranslatorInterface.php
+++ b/src/Translator/TranslatorInterface.php
@@ -6,6 +6,9 @@ use Laminas\Translator\TranslatorInterface as Translator;
 
 /**
  * Translator interface.
+ *
+ * @deprecated Since 2.27.0 The translator interface is now in a separate package `laminas/laminas-translator` and this
+ *             interface should be replaced by `Laminas\Translator\TranslatorInterface`
  */
 interface TranslatorInterface extends Translator
 {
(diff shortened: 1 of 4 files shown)


src/Translator/TranslatorInterface.php
```diff
@@ -2,36 +2,11 @@
 
 namespace Laminas\I18n\Translator;
 
+use Laminas\Translator\TranslatorInterface as Translator;
+
 /**
  * Translator interface.
```
michalbundyra, line 8: do we want to deprecate this interface in favour of `Laminas\Translator\TranslatorInterface` as well?

gsteel: Yes! Thanks @michalbundyra

boesing:
AFAIR just having an Interface extending another is not a bc break. 🤔
doctrine did this plenty of times.

asgrim:
> AFAIR just having an Interface extending another is not a bc break. 🤔 doctrine did this plenty of times.

yep, it's only a BC break if the new parent interface adds new methods that weren't there before :+1:

froschdesign (approved):

PR #141, opened 2024-10-21
Reduce use of deprecated Laminas\ServiceManager\ConfigInterface

ramchale:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | yes
| QA            | no

### Description

Laminas\ServiceManager\ConfigInterface is deprecated and removed in laminas/laminas-servicemanager v4.

Removing all uses except HelperConfig (which inherits so removal would be a breaking change) to make any future upgrade slightly easier.

### DIFF
--- a/test/Translator/TranslatorTest.php
+++ b/test/Translator/TranslatorTest.php
@@ -10,7 +10,6 @@ use Laminas\EventManager\Event;
 use Laminas\EventManager\EventInterface;
 use Laminas\I18n\Translator\TextDomain;
 use Laminas\I18n\Translator\Translator;
-use Laminas\ServiceManager\Config;
 use LaminasTest\I18n\TestCase;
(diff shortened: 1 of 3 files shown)


samsonasik (approved):

gsteel (approved):
Nice! Thanks @ramchale

test/Translator/TranslatorTest.php
```diff
@@ -157,13 +156,12 @@ public function testTranslate(): void
     {
         $loader             = new TestLoader();
         $loader->textDomain = new TextDomain(['foo' => 'bar']);
-        $config             = new Config([
+        $pm                 = $this->translator->getPluginManager();
```
froschdesign, line 159: @ramchale 
For the future: Please use meaningful names and no abbreviations like these. Also in tests.
Thanks in advance! 👍🏻

Issue #134, opened 2024-08-12
Translator factory has hidden dependency on laminas/laminas-cache-storage-deprecated-factory


rieschl:
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 2.28.0

#### Summary

We cleaned up our composer dependencies and removed the `laminas/laminas-cache-storage-deprecated-factory` because there weren't any usages in our code.
After that, the factory of the Translator exploded because it tries to instantiate the cache for the translator with the `StorageFactory` from said package.

#### Current behavior

The application explodes because `\Laminas\Cache\StorageFactory` cannot be found.

#### How to reproduce

Add
```php
    'translator' => [
        'cache' => [
            'adapter' => [
                'name' => \Laminas\Cache\Storage\Adapter\Memory::class,
            ],
        ],
    ],
```
to the config and try to pull `\Laminas\I18n\Translator\Translator` from the container or build it with `Translator::factory($options)` where `$options` is the above array inside the `translator` key.

#### Expected behavior

No boom 🙂 
I think either the dependency should be added to the `require` section and not `require-dev` or the factory should use the new `StorageAdapterFactoryInterface`(?) instead of the deprecated one.

froschdesign:
> I think either the dependency should be added to the `require` section and not `require-dev` or the factory should use the new `StorageAdapterFactoryInterface`(?) instead of the deprecated one.

The problem is the _soft_ dependency on laminas-cache. If laminas-cache is not needed then the deprecated factory is also not needed.
This will be solved with the next major version because the soft dependencies are crap. And caching in general should not be integrated in the translator or in a paginator.

ramchale:
In the meantime is it worth adding a warning to Translator?

Something like 
```
        if (isset($options['cache'])) {
            if ($options['cache'] instanceof CacheStorage) {
                $translator->setCache($options['cache']);
            } else {
                trigger_error('Constructing the cache in the Translator will be removed in a future release. Please pass a StorageInterface instance.', E_USER_DEPRECATED);
                $translator->setCache(Cache\StorageFactory::factory($options['cache']));
            }
        }
```

gsteel:
Ideally, the soft dependencies should be made hard dependencies - symbols are used in `src` therefore the dependency should be declared _(And removed in the next major)_, but, triggering a deprecation seems like a good middleground.

WDYT @froschdesign?

froschdesign:
> …triggering a deprecation seems like a good middleground.

[...]

PR #142, opened 2024-10-22
Remove dev dependency on laminas/laminas-cache-storage-deprecated-factory

ramchale:
|    Q          |   A
|-------------- | ------
| Documentation |/no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | yes
| QA            | no

### Description

laminas/laminas-cache-storage-deprecated-factory states "This factory will not provide support for laminas-cache v4+ and is only meant to be used as a temporary solution in combination with other laminas components."

Continued use is likely to make future maintenance of this package difficult

While it is used in one code path in Translator, as it is not a non-dev dependency I think it's safe to assume anyone using this functionality has it listed as one of their dependencies

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,6 @@
     "require-dev": {
         "laminas/laminas-cache": "^3.12.1",
         "laminas/laminas-cache-storage-adapter-memory": "^2.3.0",
-        "laminas/laminas-cache-storage-deprecated-factory": "^1.2",
         "laminas/laminas-coding-standard": "~2.5.0",
         "laminas/laminas-config": "^3.9.0",
         "laminas/laminas-eventmanager": "^3.13",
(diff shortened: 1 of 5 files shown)


gsteel (changes requested):
Overall, not sure about this change.
Effectively, we've just dropped a dev dependency and now have an untested code path that imports symbols not defined as hard dependencies.

The undefined deps are of course not your fault 😅, however, I'm preferring the runtime deprecation of the cache option as mentioned in #134 with the addition of new tests to cover the deprecation notice.

ramchale:
No worries. Thanks for taking a look.

I'll see how the issue conversation goes, and potentially submit a PR based on the outcome of that.

froschdesign:
> "…and is only meant to be used as a temporary solution in combination with other laminas components."

laminas-cache-storage-deprecated-factory was created precisely for this component, among others.

PR #146, opened 2025-01-09
Replace laminas-cache with psr/simple-cache

marcelthole:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | yes

### Description
laminas-cache supports PSR-6 and PSR-16 and there is no reason to depend on that laminas-cache anymore. This should also reduce the overall dependencies for updating for new PHP versions or the laminas servicemanager migration.

Question: is it fine to use only the psr-16 implementation? 

### Additional Notes
I didn't found a Issue for the Todos for a new 3.0 Version for laminas-i18n. I guess there are also some other improvementes for that version like updating the servicemanager to version 4.1 and use native property types

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,14 +34,12 @@
     "require": {
         "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0",
         "ext-intl": "*",
+        "psr/simple-cache": "^3.0",
         "laminas/laminas-servicemanager": "^3.21.0",
         "laminas/laminas-stdlib": "^3.0",
         "laminas/laminas-translator": "^1.0"
     },
     "require-dev": {
-        "laminas/laminas-cache": "^3.12.2",
-        "laminas/laminas-cache-storage-adapter-memory": "^2.3.0",
-        "laminas/laminas-cache-storage-deprecated-factory": "^1.2",
         "laminas/laminas-coding-standard": "~2.5.0",
         "laminas/laminas-config": "^3.9.0",
         "laminas/laminas-eventmanager": "^3.13.1",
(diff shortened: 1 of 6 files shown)


froschdesign:
@marcelthole 
Please see also: #120 

> laminas-cache supports PSR-6 and PSR-16 and there is no reason to depend on that laminas-cache anymore.

Correct!

> Question: is it fine to use only the psr-16 implementation?

Sure!

> I didn't found a Issue for the Todos for a new 3.0 Version for laminas-i18n.

An issue report should not be the first choice for this, milestones and projects should be.

marcelthole:
> Please see also: https://github.com/laminas/laminas-i18n/pull/120

Oh looked only in the issues and not in the open PRs :/ 

> An issue report should not be the first choice for this, milestones and projects should be.
is there an idea already to create the project?
Would be great if i could update to PHP 8.4 in one of my projects but there are still some packages that doesn't support the servicemanager v4. A roadmap would help me here a lot to create the correct PRs

gsteel:
Side note @marcelthole:

Validator 3 is done, Filter v3 is nearly there. Once filter v3 is out, the next item is probably input filter major release, then form and then view.

Because i18n may require a major release to support any/all of the above, we should avoid releasing a major here until we are sure that upgrading deps doesn't force a major release, i.e. it would be bad to release v3 here prior to view v3 only to find we've got to release a v4 for comapt with view (for example)

froschdesign:
@marcelthole @gsteel 
I will create a project and add the next steps/components to show the current planned upgrade path.

gsteel:
@marcelthole - I've gone ahead and done the bare minimum for PSR-16 in #171 as a BC break for v3.0

froschdesign:
Implemented with #195 and #198. It will be released with version 3.

froschdesign:
@marcelthole 
Thanks for bringing this up. 👍🏻

PR #150, opened 2025-08-02
Add support for laminas-filter v3 compatibility

bienvenuushindi:
**(WORK IN PROGRESS)**
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description
**(WORK IN PROGRESS)**
In version 3 of `laminas-filter` , `Laminas\Filter\AbstractFilter` has been deprecated and removed. As AbstractLocale currently extends `AbstractFilter`, it must be refactored to directly implement `FilterInterface` to ensure future compatibility and align with Laminas-filter v3 architecture decisions.

Reference: https://docs.laminas.dev/laminas-filter/v2/migration/preparing-for-v3/#abstractfilter-removal

### DIFF
--- a/src/Filter/AbstractLocale.php
+++ b/src/Filter/AbstractLocale.php
@@ -2,7 +2,7 @@
 
 namespace Laminas\I18n\Filter;
 
-use Laminas\Filter\AbstractFilter;
+use Laminas\Filter\FilterInterface;
 use Locale;
 
 /**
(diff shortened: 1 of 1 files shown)


gsteel:
Thanks for the patch @bienvenuushindi - however `filter@v3` requires `servicemanager@v4` and not all of `i18n`'s dependencies support `servicemanager@v4` yet, so it's likely to be a while before we can do the upgrade here.

bienvenuushindi:
Thank you for the clarification. I noticed the same issue when running composer update. I was attempting to upgrade both laminas-filter and laminas-validator to v3 in Tikiwiki CMS, but laminas-i18n currently requires version 2 of both, which blocks the upgrade for now.

PR #152, opened 2025-10-14
Drop support for PHP 8.1, add support for PHP 8.5

gsteel:
- Upgrades PHPUnit to 11.x
- Inlines the trait for plugin manager compatibility (Outdated and incompatible with PHPUnit 11)
- Expands baseline with laminas-view deprecations


.laminas-ci.json
```diff
@@ -1,7 +1,7 @@
 {
   "ignore_php_platform_requirements": {
-    "8.3": false,
-    "8.4": true
+    "8.4": false,
```
Xerkus: drop the 8.4. it does nothing with false

PR #154, opened 2025-12-14
Add a Makefile and installs a range of QA tools

gsteel:
Adds:

- composer require checker
- rector
- infection
- composer unused

… as local development tools, and also updates CI to run them.

Note: Infection isn't turned on in CI yet.

Adds a Makefile to run various QA checks for code and docs along with a Dockerfile with all the required extensions for this lib (i.e. intl + tooling requirements)


composer.json
```diff
@@ -33,10 +33,13 @@
...
     "require": {
         "php": "~8.2.0 || ~8.3.0 || ~8.4.0 || ~8.5.0",
+        "ext-ctype": "*",
         "ext-intl": "*",
+        "laminas/laminas-escaper": "^2.0",
         "laminas/laminas-servicemanager": "^3.21.0",
         "laminas/laminas-stdlib": "^3.0",
-        "laminas/laminas-translator": "^1.0"
+        "laminas/laminas-translator": "^1.0",
+        "psr/container": "^1.0.0"
```
Ocramius, line 42: Should we allow `^2`?

composer.json
```diff
@@ -33,10 +33,13 @@
     },
     "require": {
         "php": "~8.2.0 || ~8.3.0 || ~8.4.0 || ~8.5.0",
+        "ext-ctype": "*",
         "ext-intl": "*",
+        "laminas/laminas-escaper": "^2.0",
```
Ocramius, line 38: I'm all for explicitness, but I haven't seen this in commit messages? Therefore asking if it's an intentional addition

Ocramius (changes requested):
Commits include a baseline refresh, but I see no diff on the baseline?

composer.json
gsteel: Yes - it's intentional, and an obvious dependency here 👍

composer.json
gsteel: There's no point because SMv3.x doesn't allow it, so we can't test with it

gsteel:
> Commits include a baseline refresh, but I see no diff on the baseline?

Perhaps I missed it from the relevant commit. There is a diff on "all commits". There's a few minor fixes and some changes due to a change from `switch` to `match` which is still _very_ squishy

Ocramius (approved):
LGTM

Ocramius:
Thanks @gsteel!

PR #171, opened 2025-12-16
Swap out Laminas\Cache with Psr\Cache

gsteel:
Changes the translator to use a PSR-6 cache so that we can de-couple dependencies somewhat…

Does the bare minimum of compat work with Service Manager v4.

The translator itself still needs a big refactor…


froschdesign:
> The translator itself still needs a big refactor…

I would prefer it if we could extract the caching from the translator but for the current development, the step is fine.

froschdesign (approved):

PR #173, opened 2025-12-16
Replace Laminas\Config soft-dep by re-implementing the Ini File reader

### LINKED ISSUE #126: [RFC]: Deprecate & Remove Ini format translation files
## RFC

| BC Break?           | Yes |

### Description

`laminas-config` is no longer maintained but the ini file loader available here has a hard dependency on it.

I can't imagine that ini files are a popular format for translations, so I suggest deprecation in the 2.x series for removal in 3.0

This is also one of several dependencies that would prohibit upgrading to service-manager v4

gsteel:
Closes #126


froschdesign (approved):
LGTM 👍🏻

PR #182, opened 2025-12-17
First Pass at Refactoring the Concrete Translator

gsteel:
- Makes the loader manager a hard constructor dependency
- Add an interface for the loader manager so it can be stubbed/mocked/replaced
- Make the event manager an optional constructor dependency
- Document internal array shapes
- Add value objects to represent translation files and file patterns, improving type safety

### DIFF
--- a/src/Translator/LoaderPluginManagerInterface.php
+++ b/src/Translator/LoaderPluginManagerInterface.php
@@ -1,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\I18n\Translator;
+
+use Laminas\I18n\Translator\Loader\FileLoaderInterface;
+use Laminas\I18n\Translator\Loader\RemoteLoaderInterface;
+use Laminas\ServiceManager\PluginManagerInterface;
+
+/**
+ * @psalm-type InstanceType = RemoteLoaderInterface|FileLoaderInterface
+ * @extends PluginManagerInterface<InstanceType>
+ */
+interface LoaderPluginManagerInterface extends PluginManagerInterface
+{
+}
--- a/src/Translator/Translator.php
+++ b/src/Translator/Translator.php
@@ -10,7 +10,6 @@ use Laminas\EventManager\EventManagerInterface;
 use Laminas\I18n\Exception;
 use Laminas\I18n\Translator\Loader\FileLoaderInterface;
 use Laminas\I18n\Translator\Loader\RemoteLoaderInterface;
-use Laminas\ServiceManager\ServiceManager;
 use Laminas\Stdlib\ArrayUtils;
 use Laminas\Translator\TranslatorInterface;
 use Locale;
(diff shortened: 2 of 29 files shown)


src/Translator/LoaderPluginManagerInterface.php
```diff
@@ -0,0 +1,17 @@
...
+
+use Laminas\I18n\Translator\Loader\FileLoaderInterface;
+use Laminas\I18n\Translator\Loader\RemoteLoaderInterface;
+use Laminas\ServiceManager\PluginManagerInterface;
+
+/**
+ * @psalm-type InstanceType = RemoteLoaderInterface|FileLoaderInterface
+ * @extends PluginManagerInterface<InstanceType>
+ */
+interface LoaderPluginManagerInterface extends PluginManagerInterface
```
froschdesign: Wouldn't `MessagesLoader` be a better name with more meaning than just `Loader`?

src/Translator/Translator.php
```diff
@@ -115,7 +123,7 @@ class Translator implements TranslatorInterface
      * @return static
      * @throws Exception\InvalidArgumentException
      */
-    public static function factory($options)
+    public static function factory(MessageLoaderPluginManagerInterface $pluginManager, $options)
```
froschdesign, line 126: Will the factory method be extracted in the next steps of refactoring? That would be great, otherwise we still have the information in the translator that there are files for translations, even though the translations are already being loaded externally.

src/Translator/Translator.php
```diff
@@ -468,68 +431,70 @@ protected function getTranslatedMessage(
...
-     * @param  string $baseDir
-     * @param  string $pattern
-     * @param  string $textDomain
+     * @param non-empty-string $type
+     * @param non-empty-string $baseDir
+     * @param non-empty-string $pattern
+     * @param non-empty-string $textDomain
      * @return $this
      */
     public function addTranslationFilePattern(
```
froschdesign, line 471: The same here, can we extract this methods during the refactoring?

froschdesign:
@gsteel 
I think we should have issue reports or a project for the entire rewrite otherwise, we do not know where the journey is headed and what the individual steps are or which ones are necessary.

src/Translator/Translator.php
gsteel: Yes, I plan to extract the `factory` method to an actual factory

src/Translator/Translator.php
gsteel: I think that these methods should be removed in favour of configuration.

gsteel:
@froschdesign 
Ack! I'll try to get some time to document plans.

This patch is mainly about improving types, reducing the baseline and making sense of what's already here to make further refactorings smaller - just trying to keep the patch size down as much as possible

froschdesign (approved):
Good steps in the right direction! 👍

PR #186, opened 2025-12-20
Introduce `I18nDefaults`

gsteel:
This value object encapsulates various configuration parameters that are needed by disparate libs/classes.

Instead of parsing configuration in multiple libraries, we do it once here and provide a service that validators/filters/view helpers and users can fetch for these important defaults.


froschdesign (approved):
Nice! 👍🏼

PR #187, opened 2025-12-29
Clean Up Plural Rule Parser

gsteel:
Adds types everywhere and resolves static analysis issues by introducing a concrete type for an ast node instead of a recursive array structure.

Aggressively marks related classes as internal and readonly where appropriate


froschdesign (approved):
LGTM 👍

PR #190, opened 2025-12-31
Refactor Loader Implementations and General QA

gsteel:
Adds factories for the loaders and the ini file reader

Along with updated configuration - users can now configure behaviour of the loaders and the ini file reader via application config - for example, whether the include path should be used or not.

Cleans up tests, removing the abstract test case.

Marks more classes as `@internal`, reducing api surface.

General reduction in SA issues - down to about ~25 now.


src/ConfigProvider.php
```diff
@@ -31,6 +46,19 @@ public function __invoke(): array
...
+                    'process-sections' => true,
+                    'typed'            => false,
+                    'nest-separator'   => '.',
+                ],
+                'gettext-loader-options' => [
+                    'use-include-path' => false,
+                ],
+                'php-loader-options'     => [
+                    'use-include-path' => false,
+                ],
```
froschdesign: These are the default values, so the question arises as to whether this makes sense here.

gsteel: They are indeed unnecessary. The reason I added them is primarily for code documentation of the defaults and the expected array shape - users can look at the config provider as a template for how/what to override in their own projects.

froschdesign: I thought so, but it doesn't mention anywhere, not even that this part is actually unnecessary.

gsteel: Will adding some inline comments work for you?

froschdesign: In mezzio-laminasviewrenderer there are comments: https://github.com/mezzio/mezzio-laminasviewrenderer/blob/34b82ca7c08044bbe08d84035d9eba974a635d1d/src/ConfigProvider.php#L32-L63

froschdesign (approved):
LGTM 👍🏻

PR #176, opened 2025-12-17
Remove the laminas-mvc Module class

gsteel:

PR #192, opened 2026-01-07
Remove MVC affordances and clean up the `LoaderPluginManagerFactory`

gsteel:
Some MVC remnants remained in the `LoaderPluginManagerFactory` from #176. This patch removes them and has a cleanup of the factory.


froschdesign (approved):
Less code is always better. Go for it! 👍🏻

PR #194, opened 2026-01-08
Introduce Message Collectors to simplify the translator and externalise retrieval of translation messages

gsteel:
This is quite a big patch unfortunately, but it should be easier to follow by commit.

Using the same configuration for files, patterns and remotes as per legacy, the container can now produce `Translation Collectors` which the translator itself will consume.

There are 4 types of collector:

- File List: i.e. a list of file paths to translations
- File Patterns: zero or more patterns used to load multiple files
- Remote Services: zero or more remote services to call for fetching messages
- Aggregate: A collector that contains other collectors - used to iterate over multiple collectors in order to combine translations from multiple sources.

The aggregate collector can be configured with an explicit order so that message merging happens in a predictable order.

The translator gets a lot of code removal because of this, mainly the removal of the static factory method.


froschdesign (approved):
Good extraction! 👍🏻

PR #195, opened 2026-01-08
Add PSR6 Caching Translation Collector

gsteel:
… with a pre-wired delegator factory that conditionally decorates the aggregate collector when a cache service has been configured.

Also removes cache features from the translator

BTW, it would be trivial to support PSR-16 at the same time (in another patch)


froschdesign (approved):
Next nice step! 👍

gsteel:
@froschdesign Do you think it's worth also adding PSR-16 support?

PR #198, opened 2026-01-08
Add support for PSR-16 caches

gsteel:
Now that caching is delegated around collectors, it is trivial to support both PSR-6 and PSR-16 caches


froschdesign (approved):
Simple but effective solution with the delegator for both PSR types. 👍🏻

PR #199, opened 2026-01-09
Configurable Default Text Domain

gsteel:
Updates translator and its factory so that the configured default text domain is provided to the translator and amends references to the string literal `'default'` to use the constant from `TranslatorInterface`


froschdesign (approved):
👍🏻

PR #204, opened 2026-03-09
Replace laminas-eventmanager with PSR-14 event dispatcher

fezfez:


froschdesign:
@arueckauer 
Would you like to include this in version 3?

arueckauer:
@froschdesign Thanks for the ping. I don't personally need this change, so my preference would be to focus on getting 3.0 out the door.

froschdesign:
@arueckauer 
> …so my preference would be to focus on getting 3.0 out the door.

That’s exactly what I wanted to know. Thanks! 👍🏻

gsteel:
Thanks for your work here @fezfez, @arueckauer has re-implemented in #216 so closing here 👍

PR #212, opened 2026-06-02
Fix: Plural translations were not correctly loaded by the ini file loader.

gsteel:
The ini loader was silently rejecting lists of translation values, meaning that plural translations would not be loaded or used when using ini file format

### DIFF
--- a/test/Translator/Loader/IniTest.php
+++ b/test/Translator/Loader/IniTest.php
@@ -86,6 +86,20 @@ final class IniTest extends TestCase
         self::assertEquals('Message 4 (en)', $textDomain['Message 4']);
     }
 
+    public function testLoaderReturnsAnArrayOfEntriesForPlurals(): void
+    {
+        $loader     = new IniLoader(new IniFileReader());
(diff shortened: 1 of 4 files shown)


froschdesign (approved):
Nice! 👍🏼

test/Translator/Loader/IniTest.php
```diff
@@ -9,10 +9,12 @@
 use Laminas\I18n\Translator\Loader\IniFileReader;
 use Laminas\I18n\Translator\Plural\Rule;
 use Laminas\I18n\Translator\TextDomain;
+use Laminas\ServiceManager\ServiceManager;
 use PHPUnit\Framework\TestCase;
 
 use function realpath;
 
+/** @psalm-import-type ServiceManagerConfiguration from ServiceManager */
```
arueckauer: Is that a leftover? Could not find any usages.

gsteel: Thanks @arueckauer, well spotted!

gsteel:
Thanks to @arueckauer for the initial bug report on slack 👍

PR #216, opened 2026-06-16
Feature: PSR-14 Event Dispatcher

### LINKED ISSUE #183: Replace laminas-eventmanager with PSR-14 event dispatcher
Instead of a specific dependency on an event manager library, a PSR interface should be used.

https://www.php-fig.org/psr/psr-14/

arueckauer:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | yes
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

Based on #204 - thanks to @fezfez for all the preparation 🙏🏻 - this replaces the laminas-eventmanager with the PSR-14 event dispatcher as described in #183.

### DIFF
--- a/docs/book/v3/translator/events.md
+++ b/docs/book/v3/translator/events.md
@@ -1,16 +1,16 @@
 # Events
 
-The translator of laminas-i18n triggers two events during the processing of
+The translator of laminas-i18n dispatches two events during the processing of
 translations:
 
-* `Laminas\I18n\Translator\Translator::EVENT_MISSING_TRANSLATION`
-* `Laminas\I18n\Translator\Translator::EVENT_NO_MESSAGES_LOADED`
+- `Laminas\I18n\Translator\Event\MissingTranslationEvent`
+- `Laminas\I18n\Translator\Event\NoMessagesLoadedEvent`
 
 The typical usage for these events is to log missing translations and track when
 the loading of messages fails.
 
 > MISSING: **Installation Requirements**
-> The event support of laminas-i18n depends on the [laminas-eventmanager](https://docs.laminas.dev/laminas-eventmanager/) component, so be sure to have it installed before getting started:
+> The event support of laminas-i18n depends on a PSR-14 compatible event dispatcher.
 >
 > ```bash
 > $ composer require laminas/laminas-eventmanager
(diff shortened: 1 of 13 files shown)


src/Translator/Translator.php
```diff
@@ -47,20 +35,10 @@ public function __construct(
...
-
-        $eventManager ??= new EventManager();
-        $eventManager->setIdentifiers([
-            self::class,
-            'translator',
-        ]);
-
-        $this->events = $eventManager;
+        // When an EventDispatcher is supplied to the constructor, enable events. The user clearly wants them!
+        $this->eventsEnabled = $this->events !== null;
```
gsteel: I think we should drop this, along with enable/disable/is methods, and internally just check `$this->events !== null` - either we have a dispatcher, or we don't.

src/Translator/Translator.php
```diff
@@ -201,21 +179,14 @@ private function getTranslatedMessage(
...
+            if ($event instanceof MissingTranslationEvent) {
+                $last = $event->getTranslation();
 
-            /** @psalm-var mixed $last */
-            $last = $results->last();
-            if (is_string($last)) {
-                return $last;
+                if (is_string($last) && $last !== '') {
+                    return $last;
+                }
```
gsteel: Can we return early? i.e:

```php
if ($this->events === null) {
    return null;
}

$event = $this->events->dispatch(new MissingTranslationEvent($message, $locale, $textDomain));

if (! $event instanceof MissingTranslationEvent) {
    return null;
}

$translation = $event->getTranslation();

return is_string($translation) ? $translation : null;
```

src/Translator/Translator.php
```diff
@@ -239,20 +210,14 @@ private function loadMessages(string $textDomain, string $locale): void
...
-                if ($last instanceof TextDomain) {
-                    $messages = $last;
+            if ($this->isEventManagerEnabled() && $this->events !== null) {
+                $event = $this->events->dispatch(new NoMessagesLoadedEvent($locale, $textDomain));
+
+                if ($event instanceof NoMessagesLoadedEvent) {
+                    $last = $event->getMessages();
+                    if ($last instanceof TextDomain) {
+                        $messages = $last;
+                    }
```
gsteel: Same comment about returning early here, for eg:

```php
if ($messagesLoaded > 0) {
    $this->messages[$textDomain][$locale] = $messages;

    return;
}

if ($events === null) {
    return;
}

$event = $this->events->dispatch(new NoMessagesLoadedEvent($locale, $textDomain));

if (! $event instanceof NoMessagesLoadedEvent) {
    return;
}

$messages = $event->getMessages();
if (! $messages instanceof TextDomain) {
    return;
}

$this->messages[$textDomain][$locale] = $messages;
```

src/Translator/Event/MissingTranslationEvent.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\I18n\Translator\Event;
+
+/**
+ * Event fired when the translation for a message is missing.
+ */
+final class MissingTranslationEvent
```
gsteel: Should this event implement `StoppableEventInterface` and then prevent further execution of listeners after the first listener provides a positive result?

This means that the first successful listener wins rather than the last which I believe is how it previously worked with Laminas event manager via `triggerUntil`

src/Translator/Event/NoMessagesLoadedEvent.php
```diff
@@ -0,0 +1,51 @@
...
+declare(strict_types=1);
+
+namespace Laminas\I18n\Translator\Event;
+
+use Laminas\I18n\Translator\TextDomain;
+
+/**
+ * Event fired when no messages were loaded for a locale/text-domain combination.
+ */
+final class NoMessagesLoadedEvent
```
gsteel: Same here regarding `StoppableEventInterface`

arueckauer:
@gsteel Thanks for your thoughts. I have addressed all of them. This is now ready for a review.

docs/book/v3/translator/events.md
```diff
@@ -19,53 +19,54 @@ the loading of messages fails.
...
 
-To disable the event manager, call the `disableEventManager()` method.
+To disable the event dispatcher, call the `disableEventManager()` method.
 
 ```php
 $translator->disableEventManager();
 ```
 
-### Check Availability of Event Manager
+### Check Availability of Event Dispatcher
```
froschdesign, line 67: Can be removed.
However, we should point out that this is automatically handled simply via the availability in the DI container.
I’ll include this in my pull request, so it can stay as it is here.

gsteel (approved):
No notes! Thank you @arueckauer 👍

PR #233, opened 2026-09-15
Integrate laminas-i18n-phone-number documentation

### LINKED ISSUE #90: Add documentation for new phone number component
https://github.com/laminas/laminas-i18n-phone-number

froschdesign:
Update MkDocs navigation and add Makefile task for fetching related resources.

(This is a temporary solution to ensure that the documentation relating to the telephone numbers is displayed at all. A new documentation system must be able to resolve this directly in future.)


gsteel (commented):
Hey @froschdesign - This solves the problem nicely for now, but it requires a maintainer runs the make target and commits the changes (if any) which is easy to forget.

We could add a CI job such as `make job-name && git diff --exit-code` to fail CI if the docs are out of sync? That might be pointless though if the checkout is pinned to a specific version 🤔

froschdesign:
Another idea is to fetch the latest version based on the current major version and then check out the documentation for the phone number. This can be done as a sub-task before generating the documentation.

gsteel:
I'm wondering if we should put this on ice for a bit. One major problem here is that the phone number component is not currently installable with `i18n@3.x` because it needs a new major release to upgrade `filter`, `validator`, `i18n` and `input-filter`.

Given that `input-filter` is still not done, and, once it is, `form` will not likely be compatible with anything anymore either, so, it might make sense to add the docs to v2 (here), but v3 support is a way off.

In response to your comment and fetching the latest major, this would need GH api access wouldn't it?

froschdesign:
> I'm wondering if we should put this on ice for a bit.

Of course, after all, this is simply a starting point for discussion.

> …fetching the latest major, this would need GH api access wouldn't it?

Most likely.

